PR #1, opened 2023-03-22
requirement CRD(s)

henderiw:
This commit contains the requirement CRD(s) for nephio. These CRD(s) are used in the packages but not committed to the api server. The are used during package management and package specialization to coordinate the requirements for the network functions.


nf_requirements/v1alpha1/interface.go
```diff
@@ -0,0 +1,57 @@
...
+	Status InterfaceStatus `json:"status,omitempty" yaml:"status,omitempty"`
+}
+
+// TBD how do we distinguish the loopback from the vnic(s)
+type InterfaceSpec struct {
+	// NetworkInstance defines the networkInstance to which this interface belongs
+	NetworkInstance *references.NamespaceReference `json:"networkInstance" yaml:"networkInstance"`
+	// Cni defines the cni that is used to attach the interface to the pod
+	// +kubebuilder:validation:Enum=sriov;ipvlan;macvlan
+	Cni Cni `json:"cniType,omitempty" yaml:"cniType,omitempty"`
```
johnbelamaric, line 42: Should this be "CNI" for the Go type, ala https://github.com/containernetworking/cni/blob/spec-v1.0.0/pkg/types/types.go#L61

Or maybe this should be "CNIType"?

nf_requirements/v1alpha1/capacity.go
```diff
@@ -0,0 +1,56 @@
...
+
+type CapacitySpec struct {
+	// MaxUplinkThroughput defines the max uplink dataplane throughput
+	MaxUplinkThroughput resource.Quantity `json:"maxUplinkThroughput,omitempty" yaml:"maxUplinkThroughput,omitempty"`
+	// MaxDownlinkThroughput defines the max downlink dataplane throughput
+	MaxDownlinkThroughput resource.Quantity `json:"maxDownlinkThroughput,omitempty" yaml:"maxDownlinkThroughput,omitempty"`
+	// MaxSessions defines the max sessions of the control plane
+	MaxSessions int `json:"maxSessions,omitempty" yaml:"maxSessions,omitempty"`
+	// MaxSubscribers defines the max subscribers
+	MaxSubscribers int `json:"maxSubscribers,omitempty" yaml:"maxSubscribers,omitempty"`
```
johnbelamaric, line 44: What's the units here and for sessions? Fine if it's just singular but maybe it should be in 1000s or something?

nf_requirements/v1alpha1/dnn.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"reflect"
+
+	"github.com/nephio-project/api/references"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// +kubebuilder:object:root=true
+type DataNetworkName struct {
```
johnbelamaric, line 28: Should be be "DataNetworkName" or just "DataNetwork"?

nf_requirements/v1alpha1/dnn.go
```diff
@@ -0,0 +1,62 @@
...
+
+	Spec   DataNetworkNameSpec   `json:"spec,omitempty" yaml:"spec,omitempty"`
+	Status DataNetworkNameStatus `json:"status,omitempty" yaml:"status,omitempty"`
+}
+
+type DataNetworkNameSpec struct {
+	// Pools defines the parameters of the IP pool associated with the DNN
+	Pools []*Pool `json:"pools,omitempty"`
+	// NetworkInstance defines the networkInstance context to which this DNN belongs
+	NetworkInstance *references.NamespaceReference `json:"networkInstanceReference" yaml:"networkReference"`
```
johnbelamaric, line 40: I understand wanting to reference things outside the namespace for this, but could there be permissions issues or leaks?

nf_requirements/v1alpha1/dnn.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+type Pool struct {
+	// Name defines the name of the pool
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name,omitempty" yaml:"name,omitempty"`
+	// PrefixLength define the size of the pool
+	PrefixLength uint8 `json:"prefixLength,omitempty" yaml:"prefixLength,omitempty"`
```
johnbelamaric, line 50: Wow, so we're not going to support 512 bit addresses? :-P

nf_requirements/v1alpha1/interface.go
```diff
@@ -0,0 +1,57 @@
...
+	// AttachmentType defines if the interface is attached using a vlan or not
+	// +kubebuilder:validation:Enum=none;vlan
+	AttachmentType AttachmentType `json:"attachementType,omitempty" yaml:"attachementType,omitempty"`
```
johnbelamaric, line 45: s/attachement/attachment/g

johnbelamaric: Is there a linter that would flag that?

nf_requirements/v1alpha1/interface_types.go
```diff
@@ -0,0 +1,33 @@
...
+package v1alpha1
+
+type AttachmentType string
+
+// AttachmentTypeNone defines an untagged attachement (no VLAN)
+const AttachmentTypeNone AttachmentType = "none"
+// AttachmentTypeVLAN defines a tagged/vlan attachement
+const AttachmentTypeVLAN AttachmentType = "vlan"
+
+type Cni string
```
johnbelamaric, line 26: I think "Cni" probably should be "CNI" everywhere.

references/references.go
```diff
@@ -0,0 +1,150 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package references
+
+
+// NamespaceReference is used to reference a particular krm intstance
+type NamespaceReference struct {
```
johnbelamaric, line 21: We do have this already:

https://pkg.go.dev/k8s.io/api@v0.26.3/core/v1#ObjectReference

Is that too general?

references/references.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// Name is unique within a namespace to reference a nfClass.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
+}
+
+// PackageRevisionReference is used to reference a particular package revision.
+type PackageRevisionReference struct {
```
johnbelamaric, line 37: Not sure this should be here. We also have this: https://github.com/GoogleContainerTools/kpt/blob/porch/v0.0.15/porch/controllers/packagevariantsets/api/v1alpha1/packagevariantset_types.go#L56

references/references.go
```diff
@@ -0,0 +1,150 @@
...
+	// Revision is the specific version number of the revision of the package
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Revision string `json:"revision"`
+}
+
+// GVKReference identifies an API object containing controller-specific
+// configuration resource within the cluster.
+type GVKReference struct {
```
johnbelamaric, line 66: Are similar things available in K8s code?

johnbelamaric (changes requested):
A few comments. I am not sure that the "references" package should be here - or, at least, some of the types seem like they would just be used directly from the upstream projects if we need them. I'd prefer to add them later after we see how they are being used.

nf_requirements/v1alpha1/interface.go
johnbelamaric: See also: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions

references/references.go
henderiw: I will remove this. This is defined in Gateway API fair.

references/references.go
henderiw: ok will remove

references/references.go
henderiw: I am fine using this but we don't need most of these objects

references/references.go
johnbelamaric: Ok...though I am not *sure* we want to take a dependency on GW API. So if we need this we may want to keep it here. If multiple projects are using some similar stuff we may want to upstream it to somewhere in more core K8s.

nf_requirements/v1alpha1/interface_types.go
henderiw: agreed

nf_requirements/v1alpha1/dnn.go
henderiw: this is coming from the standard PrefixLength as ipv6 allows for prefix length 128

nf_requirements/v1alpha1/dnn.go
henderiw: it allows the network package to be actuated in a different namespace + allow for distribution to workload. We don't have to use namespace if we don't have to, but I believe it is good to have the option.

Changed it to corev1.ObjectReference which is way more generic, but I added a comment that we would only use Name and optionally namespace.

nf_requirements/v1alpha1/dnn.go
henderiw: yes good suggestion, changed it

nf_requirements/v1alpha1/interface.go
henderiw: ok I changed to CNIType again. I was going back and fwd, but lets settle on CNIType.

nf_requirements/v1alpha1/dnn.go
johnbelamaric: I know, I was joking :)

nf_requirements/v1alpha1/capacity.go
henderiw: ok I added this in the text

references/references.go
henderiw: yes, but for now I believe I can remove it as we don't need it so far.

nf_requirements/v1alpha1/interface.go
henderiw: I am not sure if the linter does this

nf_requirements/v1alpha1/interface_types.go
```diff
@@ -0,0 +1,33 @@
...
+const AttachmentTypeNone AttachmentType = "none"
+// AttachmentTypeVLAN defines a tagged/vlan attachement
+const AttachmentTypeVLAN AttachmentType = "vlan"
+
+type CNIType string
+
+// CNITypeSRIOV defines the sriov cni
+const CNITypeSRIOV CNIType = "sriov"
+// CNITypeIPVlan defines the ipvlan cni
+const CNITypeIPVlan CNIType = "ipvlan"
```
johnbelamaric, line 31: above you used VLAN, but here Vlan...

nf_requirements/v1alpha1/dnn.go
henderiw: cool

nf_requirements/v1alpha1/interface_types.go
henderiw: ok fixed it

nf_requirements/v1alpha1/interface.go
johnbelamaric: I think this is the last remaining fix needed (get rid of the extra "e" in the json and yaml tags).

henderiw: fixed

johnbelamaric:
(prow not enabled yet...)

PR #2, opened 2023-03-29
added getter api-methods

henderiw:
This PR adds Getter methods to the api objects to ease the retrieval of the information in various consumers of this library.


nf_requirements/v1alpha1/capacity_types.go
```diff
@@ -0,0 +1,35 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import "k8s.io/apimachinery/pkg/api/resource"
+
+func (r *Capacity) GetMaxUplinkThroughput() resource.Quantity {
+	return r.Spec.MaxDownlinkThroughput
```
gvbalaji, line 22: Thinking is there a merit to define an interface for these getters? That may be useful if we just need to provide Get access to the structure underneath which can be done through an interface. This is a general comment applicable for capacity, dnn and interface types in this PR

nf_requirements/v1alpha1/interface_types.go
```diff
@@ -16,18 +16,42 @@ limitations under the License.
...
 // AttachmentTypeNone defines an untagged attachement (no VLAN)
 const AttachmentTypeNone AttachmentType = "none"
+
 // AttachmentTypeVLAN defines a tagged/vlan attachement
 const AttachmentTypeVLAN AttachmentType = "vlan"
 
 type CNIType string
 
 // CNITypeSRIOV defines the sriov cni
 const CNITypeSRIOV CNIType = "sriov"
```
gvbalaji, line 34: Nitpick. Are these CNI constants defined elsewhere that we can reuse?

nf_requirements/v1alpha1/dnn_tests.go
```diff
@@ -0,0 +1,88 @@
...
+		t.Run(name, func(t *testing.T) {
+			got := tc.input.GetNetworkInstance()
+
+			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Errorf("TestGetNetworkInstanc: -want, +got:\n%s", diff)
+			}
+
+		})
+	}
+}
```
gvbalaji, line 88: Tests missing for other DNN methods such as GetPools. Simple though.

gvbalaji (changes requested):
Overall looks good. Added a suggestion and minor comments.

nf_requirements/v1alpha1/capacity_types.go
henderiw: I did something like this in the past and I did not found any benefit doing this. So my view is that it is better to use the method directly on the GoStruct. So an interface adds overhead with little value.

nf_requirements/v1alpha1/interface_types.go
henderiw: I have been looking at this indeed and e.g. in NAD these are just strings. So in my research I did not find any.

nf_requirements/v1alpha1/dnn_tests.go
henderiw: added the remaining tests

nf_requirements/v1alpha1/interface_types.go
gvbalaji: OK thanks.

gvbalaji (approved):
Thanks Wim. /lgtm

nf_requirements/v1alpha1/capacity_interfaces.go
```diff
@@ -0,0 +1,39 @@
...
+limitations under the License.
+*/
+
+package v1alpha1
+
+import "k8s.io/apimachinery/pkg/api/resource"
+
+// GetMaxUplinkThroughput return the max uplinkthroughput
+func (r *Capacity) GetMaxUplinkThroughput() resource.Quantity {
+	return r.Spec.MaxDownlinkThroughput
```
johnbelamaric, line 23: I guess I am a little confused as to the purpose of the libraries. Typically in Go we don't create getters for exported fields (capitalized).

I thought we wanted Getters for when the underlying resources we were dealing with were kyaml (or similar, yaml manipulation). If we already have it parsed into a Go object, can't we just read the field directly?

Makefile
```diff
@@ -37,6 +37,18 @@ help: ## Display this help.
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
+.PHONY: fmt
```
vjayaramrh, line 40: You may want to consider using the Makefile contents from https://github.com/nephio-project/nephio/blob/main/Makefile

nf_requirements/v1alpha1/capacity_interfaces.go
henderiw: You can indeed q the value of this here. Some people avoid processing through direct struct and the value is when you have a slice or map to help deal with insertion/deletion. Right now we don't have these complex structs so these are 1:1 mapping. But if we have more complex structure we would add some getter/setters. So the idea was to generalise this.

Makefile
henderiw: added the new makefile

PR #4, opened 2023-04-03
Add OWNERS file

johnbelamaric:


johnbelamaric:
@radoslawc is Prow configured with the right org? Or is it pointing at K8s, perhaps?

PR #6, opened 2023-04-04
Fix OWNERS

johnbelamaric:
/assign @radoslawc


johnbelamaric:
/close

race condition

PR #8, opened 2023-04-07
UPFDeployment definition check-in

denysaleksandrov:
Adding  UPFDeployment type and yaml definition files:
- config/crd/bases/req.nephio.org_upfdeployments.yaml
- nf_requirements/v1alpha1/upfdeployment_types.go

The definition is based on CRD used during ONE Summit. I will move gradually to the definition in this [doc](https://docs.google.com/document/d/1cD-Lo-isvi1vN-G_TimPqIEU_eXp2H3a9vUIaAyH6Ug/edit#heading=h.e3i35mj5jyac).
If I change it completely now, it will break the code Stephen and I wrote for free5gc upf.

Statuses defined according to this [doc](https://docs.google.com/document/d/1cD-Lo-isvi1vN-G_TimPqIEU_eXp2H3a9vUIaAyH6Ug/edit#heading=h.e3i35mj5jyac).


nf_requirements/v1alpha1/upfdeployment_types.go
```diff
@@ -0,0 +1,132 @@
+/*
+Copyright 2022 Authors of Project Nephio.
```
johnbelamaric, line 2: 2023 The Nephio Authors

I guess we can standardize on that one? @vjayaramrh @henderiw @gvbalaji @aravind254 @radoslawc

nf_requirements/v1alpha1/upfdeployment_types.go
johnbelamaric, line 1: This is the wrong directory, this probably belongs in something like "nf_deployments"? @tliron @s3wong @henderiw WDYT? There should be a module just for the deployment / workload cluster resources.

johnbelamaric (changes requested):

nf_requirements/v1alpha1/upfdeployment_types.go
denysaleksandrov: Changed.

denysaleksandrov:
those presubmit failed jobs - is the issue in the code or in the prow config?

denysaleksandrov:
there are no vet/fmt/test rules in Makefile in this project. should we add them?

johnbelamaric:
> `make: *** No rule to make target 'fmt'.  Stop.`

Well, I guess I would say prow config, as the errors seem to be that the Makefile doesn't have the targets that the prow config is expecting. I'll push a PR that fixes that; after it merges you can rebase.

johnbelamaric:
https://github.com/nephio-project/api/pull/9

nf_requirements/v1alpha1/upfdeployment_types.go
henderiw: yes agreed nf_deployments seems ok.

s3wong: Yes, nf_deployments seems like a more appropriate place

s3wong:
The EasyCLA failure is a bit of a mystery; the first failure seems legit; the other two are related to access (storage) related errors...

nf_requirements/v1alpha1/upfdeployment_types.go
denysaleksandrov: moved upfdeployment_types.go into nf_deployments/v1alpha1 folder.

aakashchan:
> Adding UPFDeployment type and yaml definition files:
> 
> * config/crd/bases/req.nephio.org_upfdeployments.yaml
> * nf_requirements/v1alpha1/upfdeployment_types.go
> 
> The definition is based on CRD used during ONE Summit. I will move gradually to the definition in this [doc](https://docs.google.com/document/d/1cD-Lo-isvi1vN-G_TimPqIEU_eXp2H3a9vUIaAyH6Ug/edit#heading=h.e3i35mj5jyac). If I change it completely now, it will break the code Stephen and I wrote for free5gc upf.
> 
> Statuses defined according to this [doc](https://docs.google.com/document/d/1cD-Lo-isvi1vN-G_TimPqIEU_eXp2H3a9vUIaAyH6Ug/edit#heading=h.e3i35mj5jyac).

Package specilization kpt-function which generates `UPFDeployment` also relies on this api schema. How are planning to manage revisions on this schemas?

denysaleksandrov:
Closing. Wim is providing correct apis in https://github.com/nephio-project/api/pull/11

PR #10, opened 2023-04-11
Validate fn added to support the interface lib

henderiw:
Added a set of function to add the validation of the interface objects. Used in the interface lib as part of the arm functions


lcov.info
```diff
@@ -0,0 +1,18 @@
+mode: set
```
vjayaramrh, line 1: What is the purpose of this file?

nf_requirements/v1alpha1/interface_interfaces.go
```diff
@@ -0,0 +1,89 @@
+/*
+Copyright 2023 Nephio.
```
johnbelamaric, line 2: nit: The Nephio Authors

nf_requirements/v1alpha1/interface_types.go
```diff
@@ -16,21 +16,43 @@ limitations under the License.
...
-const CNITypeMACVLAN CNIType = "macvlan"
+import (
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// +kubebuilder:object:root=true
```
johnbelamaric, line 27: Curious: are we generating actual CRDs/Open API schemas for these?

johnbelamaric (commented):
On minor nit, otherwise LGTM

lcov.info
henderiw: sorry should not have been there. removed.

nf_requirements/v1alpha1/interface_interfaces.go
henderiw: I changed on all the files as this was not there originally.

nf_requirements/v1alpha1/interface_types.go
henderiw: it was there. Should we remove it. We don't apply to the apiserver but we can use it later for some validations if required.

nf_requirements/v1alpha1/interface_interfaces.go
```diff
@@ -0,0 +1,89 @@
...
+		return false
+	}
+	return true
+}
+
+func ValidateInterfaceSpec(spec *InterfaceSpec) error {
+	if spec == nil {
+		return fmt.Errorf("spec invalid: %s", errMissingNetworkInstance)
+	}
+	if spec.AttachmentType != "" {
```
vjayaramrh, line 72: question: Is this the case, where the AttachmentType is not set, it defaults to an empty string?

nf_requirements/v1alpha1/interface_types.go
```diff
@@ -16,21 +16,43 @@ limitations under the License.
...
+	NetworkInstance *corev1.ObjectReference `json:"networkInstance" yaml:"networkInstance"`
+	// CNIType defines the cniType that is used to attach the interface to the pod
+	// +kubebuilder:validation:Enum=sriov;ipvlan;macvlan
+	CNIType CNIType `json:"cniType,omitempty" yaml:"cniType,omitempty"`
+	// AttachmentType defines if the interface is attached using a vlan or not
+	// +kubebuilder:validation:Enum=none;vlan
+	AttachmentType AttachmentType `json:"attachmentType,omitempty" yaml:"attachmentType,omitempty"`
+}
+
+type InterfaceStatus struct {
```
vjayaramrh, line 49: curious about why the struct is empty? Thanks

nf_requirements/v1alpha1/interface_interfaces.go
```diff
@@ -0,0 +1,89 @@
+/*
```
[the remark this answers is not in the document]
henderiw: yes my scheme is the following _types.go is where the type definition of the CR go and _interfaces is where the methods go that apply to the interface.

nf_requirements/v1alpha1/interface_interfaces.go
henderiw: yes this is the default go type behaviour. There is another PR on the api repo to change these to pointers. I am ok either way, but we started without pointers.

nf_requirements/v1alpha1/interface_types.go
henderiw: so far we had no reason to add something in the status and it could be empty. Now with the new fn-runtime or sdk proposal we will add information here going fwd. So it will be used to indicate the IP(s) that were allocated e.g.
As such the nf-deploy function will not have to look to intermediate CR(s) to understand the IP.

PR #21, opened 2023-05-03
Updated cluster context with validate function

ganchandrasekaran:


infra/v1alpha1/cluster_context.go
```diff
@@ -64,3 +75,29 @@ type ClusterContextList struct {
 func init() {
 	SchemeBuilder.Register(&ClusterContext{}, &ClusterContextList{})
 }
+
+func ValidateClusterContextSpec(spec *ClusterContextSpec) error {
```
kispaljr, line 79: you may consider turning this into a method:
`func (spec ClusterContextSpec) ValidateClusterContextSpec() error` , or
`func (spec ClusterContext) ValidateClusterContextSpec() error`
but I'm not pushing, it is good as it is, as well

ganchandrasekaran: I wanted the same, but then wanted to align with interface-api 
https://github.com/nephio-project/api/blob/main/nf_requirements/v1alpha1/interface_interfaces.go#L68

infra/v1alpha1/cluster_context.go
```diff
@@ -64,3 +75,29 @@ type ClusterContextList struct {
...
+	if spec == nil {
+		return fmt.Errorf("spec invalid: %s", errMissingClusterContextSpec)
+	}
+	if *spec.SiteCode == "" {
```
kispaljr, line 83: check if spec.SiteCode was nil, before dereferencing it with *

PR #22, opened 2023-05-07
Add NFDeployment interface to support the NF Deployment KRM function 

kispaljr:
Add an NFDeployment interface, that allows NF vendors (by implementing the interface) to create SMFDeployment-like CRDs to their own NFs and then use @aakashchan's SDK to create an NFDeployment KRM function based on his code in this PR: https://github.com/nephio-project/nephio/pull/137

### DIFF
--- a/nf_deployments/v1alpha1/amf_deployment_types.go
+++ b/nf_deployments/v1alpha1/amf_deployment_types.go
@@ -48,6 +48,21 @@ type AMFDeploymentStatus struct {
 	NFDeploymentStatus `json:",inline" yaml:",inline"`
 }
 
+// Implement NFDeployment interface
+
+func (d *AMFDeployment) GetNFDeploymentSpec() *NFDeploymentSpec {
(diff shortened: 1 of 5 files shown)


aakashchan:
@henderiw @s3wong @gvbalaji Please take a look into this.

aakashchan (approved):

henderiw:
I don't see this as an multivendor implementation but a generic interface across the NF types.

Btw as a convention we put all methods in _interface.go and the type definition in _types.go. I believe this is a clean convention.

I am ok with this approach

kispaljr:
> I don't see this as an multivendor implementation but a generic interface across the NF types.
> 
> Btw as a convention we put all methods in _interface.go and the type definition in _types.go. I believe this is a clean convention.
> 
> I am ok with this approach

Moved the NFDeployment interface into an _interface.go file

gvbalaji (approved):
/lgtm

PR #23, opened 2023-05-10
repository api(s)

henderiw:
This is the repository api to creates repos


config/crd/bases/infra.nephio.org_clustercontexts.yaml
johnbelamaric, line 1: Was this just never committed before?

config/crd/bases/req.nephio.org_datanetworks.yaml
johnbelamaric, line 1: Did we just miss a generate so this was out of date?

infra/v1alpha1/repository_interfaces.go
johnbelamaric, line 1: So, is this to be used to *provision* repositories, or is this used to *specify a package's requirements* for a repository? I prefer to decouple those things. I also think we may not need the latter (or, rather, I may want to handle it as a generic package dependency instead, which we haven't figure out yet).

If it's for *provisioning*, I would think it belongs in an API group that goes with the repo provisioning controller, rather than in the general "infra" group. Just like we have talked about providing different packages for different cluster providers, we could provide different packages for different repo providers. This would be instead of abstracting the repo concept and having different implementors.

All that said, I am pretty much OK with just getting something working in R1; we can debate the details later.

config/crd/bases/infra.nephio.org_clustercontexts.yaml
henderiw: seems so as I did a generate.

infra/v1alpha1/repository_interfaces.go
henderiw: It is a bit if a mixed scenario right now and the controller could be implemented as a specialised or a real controller. The main difference I see is the feedback loop is different. Right now we decided to work as a controller so it will be used to provision a repo.
I agree we should look at this after R1 and maybe restructure things based on the patterns we see fit best for these scenario's.

config/crd/bases/req.nephio.org_datanetworks.yaml
henderiw: it was out of date it seems. We might need to integrate this as a prow job.

PR #25, opened 2023-05-11
Update the CRDs with dnsNameserver

arora-sagar:
As discussed in Tuesday's meeting. It will be good to allow users to define the dnsNameservers for DNNs. There is no urgency for this for R1 because for R1 we can still configure it via the SMF Operator. So when the user will deploy the SMF Operator they can configure the DNS nameserver which corresponds to their network. 

PS: I only changed in workload CRDs, if I have to change in the request CRDs please let me know.


henderiw (changes requested):
we should be updating the go types as these are generated from the go types

arora-sagar:
I will reopen it later when I will change the CRD with 3GPP specific parameters

PR #29, opened 2023-05-16
Downgrade K8s API-related go modules to v0.26.1 

kispaljr:
... >= v0.27.0 versions of go modules `k8s.io/api`, `k8s.io/apimachinery`, `k8s.io/client-go` are incompatible with the (currently) latest version of `controller-runtime`.
This PR downgrades those packages to avoid this problem.


gvbalaji (approved):
/approve

PR #33, opened 2023-05-26
Network CRD

henderiw:
CRD that defines the high level network intent.
It allows for bridge domains and routing tables based on the functionality supported in nephio R1.


henderiw:
close as already merged

PR #35, opened 2023-06-01
NFTopology and related CRDs initial commit

s3wong:


nf_deployments/v1alpha1/nf_deployed.go
```diff
@@ -0,0 +1,90 @@
...
+
+type NFDeployedConnectivity struct {
+	// peer NF's Id (see NFInstance struct below)
+	NeighborName string `json:"neighborName,omitempty" yaml:"neighborName,omitempty"`
+}
+
+// NFDeployedInstance defines an NF instance that is deployed
+type NFDeployedInstance struct {
+	// unique ID for this NF instance
+	Id string `json:"id,omitempty" yaml:"id,omitempty"`
```
johnbelamaric, line 30: s/Id/ID/

Also, for the ones below this, the json/yaml names are not following the conventions:
s/clustername/clusterName/
s/nftype/nfType/
s/nfflavor/nfFlavor/
etc.

nf_deployments/v1alpha1/nf_deployed.go
johnbelamaric, line 1: I don't think this name follows conventions. "Deployed" implies a past action having been done. Is this describing tings that are *disovered* - and thus should all be "status", or is this describing user intent, in which case it should be "spec" as it is, but it should be a noun. How is it conceptually different than "NFDeployment"? It seems to me it is a linkage between a cluster and a NF. Maybe the name could reflect that?

I also think it probably belongs in its own group. I think you probably need a topology.nephio.org group or something.

What about a name like "NFClusterAssignment" or something to reflect that it relates an NF to a specific cluster?

johnbelamaric (changes requested):
A few comments. The biggest thing is that I don't think this belongs in these groups, and I think we need a name for NFDeployed that follows the conventions and also better captures what the resource represents.

I think we need a separate "topology" group, and most (if not all - I haven't thought each through in detail) of these would live in there.

nf_requirements/v1alpha1/nf_topology_types.go
```diff
@@ -0,0 +1,100 @@
...
+type NFType string
+
+const (
+	NFTypeUPF NFType = "upf"
+	NFTypeSMF NFType = "smf"
+	NFTypeAMF NFType = "amf"
+)
+
+// NFAttachment defines the specification of network attachment points of a NF
+type NFAttachment struct {
```
henderiw, line 35: we changed this to interface in the blueprints as people were confused with NADs.

nf_requirements/v1alpha1/nf_topology_types.go
```diff
@@ -0,0 +1,100 @@
...
+	NFType NFType `json:"nfType,omitempty" yaml:"nfType,omitempty"`
+
+	// ClassName --- for now, the NFClass this NF template will derive from
+	ClassName string `json:"className,omitempty" yaml:"className,omitempty"`
+
+	// Capacity specifies the NF capacity profile for this NF instance
+	Capacity CapacitySpec `json:"capacity,omitempty" yaml:"capacity,omitempty"`
+
+	// NFAttachments
+	NFAttachments []NFAttachment `json:"nfAttachments,omitempty" yaml:"nfAttachments,omitempty"`
```
henderiw, line 56: same comment. propose to change to interface as this makes it consistent

nf_requirements/v1alpha1/nf_topology_types.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type NFInstance struct {
+	// Name specifies the name of this NFInstance
+	Name string `json:"name,omitempty" yaml:"name,omitempty"`
+
+	// ClusterSelector specifies the matching labels for the NF instance to be instantiated
+	ClusterSelector metav1.LabelSelector `json:"clusterSelector,omitempty" yaml:"clusterSelector,omitempty"`
```
henderiw, line 64: if this is optional it should be a pointer

nf_requirements/v1alpha1/nf_topology_types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// NFTemplate specifies the template of the NF to be deployed when a cluster matches
+	// the selector above
+	NFTemplate NFTemplate `json:"nfTemplate,omitempty" yaml:"nfTemplate,omitempty"`
```
henderiw, line 68: if this is optional it should be a pointer

nf_deployments/v1alpha1/nf_deployed.go
```diff
@@ -0,0 +1,90 @@
...
+	NFVendor string `json:"nfvendor,omitempty" yaml:"nfvendor,omitempty"`
+	// the software version of this NF vendor's NFType
+	NFVersion string `json:"nfversion,omitempty" yaml:"nfversion,omitempty"`
+	// list of connected NF instances to this NF instance
+	Connectivities []NFDeployedConnectivity `json:"connectivities,omitempty" yaml:"connectivities,omitempty"`
+}
+
+// NFDeployedSpec defines list of deployed NF for Edge Status Aggregator to monitor
+type NFDeployedSpec struct {
+	NFInstances []NFDeployedInstance `json:"nfinstances,omitempty" yaml:"nfinstances,omitempty"`
```
henderiw, line 47: Why is this not part of the status of the topology?

henderiw (changes requested):
I believe this CR does not belong in req. We could maybe do nf_service or so?
I would put the deployed within the status of the CR

s3wong:
> I don't think this name follows conventions. "Deployed" implies a past action having been done. Is this 
> describing tings that are disovered - and thus should all be "status", or is this describing user intent, in 
> which case it should be "spec" as it is, but it should be a noun. How is it conceptually different than 
> "NFDeployment"? It seems to me it is a linkage between a cluster and a NF. Maybe the name could reflect > that?

I think there is a confusion on NFDeployed (poor name, I agree). NFDeployed here is meant as a spec to drive ESA to start monitoring what it is supposed to monitor (AMF/SMF/UPFDeployment statuses for cluster A,B,C...etc). As ESA is not undergoing much changes from seed, the input CR also isn't expected to change a lot --- it is basically the same as the old "NFDeploy" CRD that was seeded before (recall the input processing piece of ESA comes from the "deployment" entity of NFDeploy controller)

As for status of NFTopology controller, I agree, it is basically reflected by seeing which packagerevision is created for this particular topology. But I haven't defined NFTopology's statuses yet (hence not part of this CRD)

s3wong:
@henderiw 
> I believe this CR does not belong in req. We could maybe do nf_service or so? I would put the deployed 
> within the status of the CR

nf_service or as John said, topology.nephio.org should be fine. Please see my reply to John on why creating a separate "NFDeployed" --- I can change the name, but that spec is meant for ESA, not for tracking status of NFTopology

nf_requirements/v1alpha1/nf_topology_types.go
s3wong: OK --- will change

nf_requirements/v1alpha1/nf_topology_types.go
s3wong: Sounds good. Will change

nf_requirements/v1alpha1/nf_topology_types.go
s3wong: This is required, not optional. One of the things I did for simplifying implementation is that the NFInstance is ONLY key'ed off of cluster selector (instead of other potential selector also). I think it should suffice for R1, and can evolve the CRD as we discover other use cases to instantiate NF instances dynamically.

nf_requirements/v1alpha1/nf_topology_types.go
s3wong: NFTemplate is definitely required. An NFInstance cannot be instantiated if there isn't a NFTemplate... this is akin to having a podspec template for Deployment, it is mandatory

nf_deployments/v1alpha1/nf_deployed.go
s3wong: See the other comments. This is meant to drive ESA to start monitoring. As ESA is undergoing very little changes, we are keeping the CRD to start status collection the same as seed.

In a more conceptual view. The NFTopology CRD is meant to specify that for a cluster matching a label, a NFTemplate needs to be deployed onto that cluster, and the status fields there are meant to track the number of NF instances expected to be created as part of this deployment, and whether they are "deployed".

ESA CRD (which is what NFDeployed is meant for) is supposed to get a list of deployment that has happened (or expected to happen), and therefore it would start collecting statuses from the target cluster where this NF is expected to be deployed.

nf_requirements/v1alpha1/nf_topology_types.go
henderiw: ok we should remove omit empty in this case

nf_deployments/v1alpha1/nf_deployed.go
henderiw: thx for clarification. Would we not reflect the result in the topology CR?

nf_requirements/v1alpha1/nf_topology_types.go
s3wong: Good point. Will do

nf_deployments/v1alpha1/nf_deployed.go
s3wong: Meaning the (aggregation of) AMF/SMF/UPFDeployment statuses should be reflected on NFTopology CR? Probably not --- based on the CRD, there are basically two stages: the issuer of NFTopology has no prior knowledge of which NF will be deployed on what cluster, and therefore the controller for this CRD should track which NF is supposed to be deploying on which cluster. Typically, without aggregation logic, the expectation is for the user to read the status of NFTopology, and kubectl to each of the target workload cluster and check each deployed NF's status.

nf_requirements/v1alpha1/nf_topology_types.go
henderiw: understood. We should keep the label selected generic but only use it for cluster-name. We did the same in various functions.

nf_deployments/v1alpha1/nf_deployed.go
henderiw: ok here is my simplistic view.
Is topology not the overall controller?
if so I assume it drives nddeploy, not?
Is so nf deploy should now the origin of the stats and as such nf topology can get the aggregate back even w/o knowing how many got deployed.

Maybe too simplistic as I have not looked at the implementation, but this is what I would do.

nf_deployments/v1alpha1/nf_deployed.go
s3wong: Changed to NFTopoTracker to reflect that the CR is to basically configure ESA to start tracking

johnbelamaric (commented):
/approve
/lgtm

PR #36, opened 2023-06-01
Add WorkloadCluster

johnbelamaric:


johnbelamaric:
This is used as an injected resource during workload cluster provisioning. Since it is injected, it needs to live in the cluster, and thus needs a real CRD. The actual Go type is not used anywhere yet, but could be at some point.

henderiw:
should we replace cluster context with this?
In cluster context we have CNI/masterinterface and site ID.

it seems odd we have 2 CRDs for a cluster. Even we don't get feedback from the cluster status at this stage, I would propose to merge them into one, as it eases the automation.

johnbelamaric:
Hmm. Maybe, but I see the cluster context as something that is controlled by the author of the cluster package. For example, it would be the creator of the capi-cluster-kind package that knows what the CNI master interface is going to be for that cluster, so they would include that in there. I agree the "clusterName" or "siteCode" aspect doesn't belong in the same place. That CRD was designed from the point of view of the function deployment, whereas this one is from the point of view of the cluster deployment.

I guess that's the question - where do the different bits of info for the cluster context come from? I think clusterName comes from here - CNI info comes from that other package. So, we can just keep them separate and inject the specific ones we need into the NF packages, or we can combine them into something - but ultimately if we combine them and say inject this CR in the NF packages, we will have to populate the CNI info into this from something in the other package, which may be tricky.

Sorry if this is rambling a bit...I am a bit tired :)

henderiw:
My 2 cents We need a resource that defines the workload cluster Intent. In that intent you specify the requirements like I want this CNI (could be adjacent to the CR but jus conceptually), etc. Even and this is out of context the provider that is going to actuate the workload cluster (right now we assume this to be CAPI, but this has to become generic).
We want this CR to come part of the package -> I see this as the binding CR going fwd.
This one is used to represent the name through the injection/binding process as well as exposes the requirements we want the workload clusters to fulfil. 
This CR comes from the author and gets consumed by the various actors.

henderiw:
Anyhow lets move fwd with this and we can discuss later how to handle this going fwd.

/approve
/lgtm

johnbelamaric:
> My 2 cents We need a resource that defines the workload cluster Intent. In that intent you specify the requirements like I want this CNI (could be adjacent to the CR but jus conceptually), etc. Even and this is out of context the provider that is going to actuate the workload cluster (right now we assume this to be CAPI, but this has to become generic). We want this CR to come part of the package -> I see this as the binding CR going fwd. This one is used to represent the name through the injection/binding process as well as exposes the requirements we want the workload clusters to fulfil. This CR comes from the author and gets consumed by the various actors.

Ok, got it. So, I was thinking of the CNI as something the lower layer advertises support for, and that we need a CRD for that. We also need a CRD for the intent author to say "I need a cluster with XYZ CNI". So probably I should have put this in "req" not "infra" to reflect that this is "intent", and then a CNI would make sense (at least, the CNI type, not so sure about CNI master interface, that seems more "demographic"). But as you said, let's work on that for R2.

PR #38, opened 2023-06-04
updated workload cluster validation

henderiw:
added additional validations for the workload cluster now that master interface and CNI are in there>
This is used in the KRM functions


johnbelamaric (approved):
/approve

PR #39, opened 2023-06-05
Interface addressing

henderiw:
Right now we only allocate 1 IP address family.
With this PR the author of the package is able to select what type of addressing he/she wants: ipv4only, ipv6only dual stack or none (which means L2)


config/crd/bases/req.nephio.org_interfaces.yaml
```diff
@@ -34,6 +34,14 @@ spec:
           spec:
             description: TBD how do we distinguish the loopback from the vnic(s)
             properties:
+              addressing:
```
johnbelamaric, line 37: K8s uses ipFamilyPolicy and ipFamilies for Service: https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec

Can/should we follow the same practice?

henderiw: I aligned with the IPFamily convention. The IPFamily Policy is slightly different, but I reused the naming

PR #40, opened 2023-06-05
Adding copyright headers

radoslawc:


henderiw:
Are we sure they will stick as they are autogenerated ?

PR #46, opened 2023-06-12
Netwcfg API

henderiw:
provides config CRD to configure network devices
provide additional configuration to the network API. Parameters that are not necessarily exposed to the intent creator. These are params for the engineers


johnbelamaric:
There seem to be some other commits mixed in. What's the new part, is it possible to split out that?

The networks CRD for example we already have, also those "cond" changes look familiar, I think they are already merged.

henderiw:
this is new.

1. the config CRD is the CRD that is the one providing the config of the network device
2. the networkConfig is a CRD that is for the network people deploying the networks. They have default for various scenario's. Like pick prefix length x is it is an internal link, etc.

PR #47, opened 2023-06-14
Initial commit for ConfigRef CRD

s3wong:


henderiw:
should we not call this ref iso nf_references as it can be generic and not specific to a NF.

nf_references/v1alpha1/nf_config_reference.go
```diff
@@ -0,0 +1,62 @@
...
+	Group string `json:"group,omitempty" yaml:"group,omitempty"`
+
+	// Version specifies the version for this ref
+	Version string `json:"version,omitempty" yaml:"version,omitempty"`
+
+	// Kind specifies the resource kind for this ref
+	Kind string `json:"kind,omitempty" yaml:"kind,omitempty"`
+
+	// Config specifies the resource content (spec) in JSON string format for this ref
+	Config string `json:"config,omitempty" yaml:"config,omitempty"`
```
henderiw, line 35: runtime.Rawextension gives us this out of the box though the object reference.

henderiw:
When in the plane I looked at this in more detail and there is a way more optimal mode to encode this. here is an example. We can embed any KRM resource in here and derive the kind/apiversion from there.

// ConfigSpec defines the desired state of the  Configuration
type ConfigSpec struct {
	// Config defines the configuration data, which is a json encapsulated resource
	//+kubebuilder:pruning:PreserveUnknownFields
	Config runtime.RawExtension `json:"config" yaml:"config"`
}

config/crd/bases/ref.nephio.org_configref.yaml
```diff
@@ -0,0 +1,60 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.9.2
+  creationTimestamp: null
+  name: configref.ref.nephio.org
```
henderiw, line 8: would it not be better to make this config iso configref since we are already in a ref namespace.

henderiw (changes requested):

s3wong:
> should we not call this ref iso nf_references as it can be generic and not specific to a NF.

Fixed. Done

nf_references/v1alpha1/nf_config_reference.go
s3wong: Done

config/crd/bases/ref.nephio.org_configref.yaml
s3wong: Done

henderiw:
@s3wong  lets remove this file.

https://github.com/nephio-project/api/pull/47/files#diff-4f3257274db72312fce01418265342f049c95e98faed39413c93d3e375147f76

s3wong:
> @s3wong lets remove this file.
> 
> https://github.com/nephio-project/api/pull/47/files#diff-4f3257274db72312fce01418265342f049c95e98faed39413c93d3e375147f76

Oops, missed that. Done. Thanks for spotting it.

PR #51, opened 2023-09-27
Reduce hardcoding of NF parameters (https://github.com/nephio-project…

VamshiKrishnaNemalikonda:
Reduce hardcoding of NF parameters (https://github.com/nephio-project/nephio/issues/340)


nf_deployments/v1alpha1/nf_deployment_types.go
```diff
@@ -34,153 +57,176 @@ type NFDeploymentSpec struct {
...
+type ObjectReference struct {
+	// APIVersion of the target resources
+	APIVersion string `yaml:"apiVersion,omitempty" json:"apiVersion,omitempty"`
+
+	// Kind of the target resources
+	Kind string `yaml:"kind,omitempty" json:"kind,omitempty"`
+
+	// Name of the target resource
+	// +optional
+	Name *string `yaml:"name" json:"name"`
```
josephthaliath, line 72: Is there a reason for making it *string instead of string

VamshiKrishnaNemalikonda: No I don't any particular reason as of now.

henderiw: The rule we follow in the api is the reverse of what we have done here.

omit empty -> means it is optional and we make it a pointer -> so *string
if mandatory it should be a string.

I believe for us all of these parameters are mandatory

nf_deployments/v1alpha1/nf_deployment_types.go
```diff
@@ -34,153 +57,176 @@ type NFDeploymentSpec struct {
 	NetworkInstances []NetworkInstance `json:"networkInstances,omitempty" yaml:"networkInstances,omitempty"`
 	// configRef defines addiitonal configuration references the nf depends upon
 	// +optional
-	ConfigRefs []corev1.ObjectReference `json:"configRefs,omitempty" yaml:"configRefs,omitempty"`
+	ParametersRefs []ObjectReference `json:"parametersRefs,omitempty" yaml:"parametersRefs,omitempty"`
+}
+
+type ObjectReference struct {
+	// APIVersion of the target resources
+	APIVersion string `yaml:"apiVersion,omitempty" json:"apiVersion,omitempty"`
```
henderiw, line 65: apiversion, kind and name should be mandatory in my view

nf_deployments/v1alpha1/nf_deployment_types.go
josephthaliath: Thanks Wim for the clarification. I also think these should be mandatory. The kind will always be Config right? the embedded struct inside spec.config will be a type that is known within the controller. But if there are multiple Config and different embedded structures, how can we map it to right structures. Should the embedded struct always have a key 'kind' which will be used for this purpose. Or do we add this info in ParametersRef (like embedded kind)

nf_deployments/v1alpha1/nf_deployment_types.go
VamshiKrishnaNemalikonda: I've modified as suggested @henderiw @josephthaliath

nf_deployments/v1alpha1/nf_deployment_types.go
henderiw: No this kind is the kind of the KRM resource we reference. So e.g. if we define 3gpp.org/v1alpha1.XXX name=foo

the kind will be XXX apiVersion: 3gpp.org/v1alpha1 and name foo

henderiw:
prow /retest

config/crd/bases/workload.nephio.org_nfdeployments.yaml
```diff
@@ -314,8 +251,28 @@ spec:
                   - name
                   type: object
                 type: array
+              parametersRefs:
+                description: configRef defines addiitonal configuration references
```
efiacor, line 255: "parametersRef defines additional"

henderiw: can you approve this PR https://github.com/nephio-project/api/pull/53
it addresses your comment

PR #55, opened 2023-10-23
Simplify NFTopology CRD

s3wong:
Changes aimed to simplify NFTopology CRD:

1. no longer has a defined NFType (following how Nephio's core NFDeployment is evolving)
2. no longer directly sends CR to ESA, instead tracks deployed NFs in its status fields
3. identify some fields that we expect to (eventually) be embedded in a CR in all NF packages (part of SDK effort) --- mainly vendor and NF software version info

--- updated kubebuilder version


config/crd/bases/topology.nephio.org_nftopologies.yaml
```diff
@@ -165,6 +186,48 @@ spec:
...
+                        eventually be metadata embedded in a NF package (NFType should
+                        be as well) // NF vendor name NFVendor string `json:\"nfVendor,omitempty\"
+                        yaml:\"nfVendor,omitempty\"` // the software version of this
+                        NF vendor's NFType NFVersion string `json:\"nfVersion,omitempty\"
+                        yaml:\"nfVersion,omitempty\"` \n list of connected NF instances
+                        to this NF instance"
+                      items:
+                        properties:
+                          neighborName:
+                            description: peer NF's Id (see NFInstance struct below)
```
efiacor, line 215: Is this needed here?
(see NFInstance struct below)

config/crd/bases/topology.nephio.org_nftopologies.yaml
```diff
@@ -165,6 +186,48 @@ spec:
...
+                description: Detail on the deployed instances.
+                items:
+                  description: NFDeployedInstance defines an NF instance that is deployed
+                  properties:
+                    clusterName:
+                      description: name of workload cluster where the NF instance
+                        is to be deployed
+                      type: string
+                    connectivities:
+                      description: "TODO(s3wong): the following parameters should
```
efiacor, line 205: Is there a better way to track this rather then a TODO?

config/crd/bases/topology.nephio.org_nftopologies.yaml
```diff
@@ -165,6 +186,48 @@ spec:
...
+                          neighborName:
+                            description: peer NF's Id (see NFInstance struct below)
+                            type: string
+                        type: object
+                      type: array
+                    id:
+                      description: unique ID for this NF instance
+                      type: string
+                    nfType:
+                      description: 'type of NF, example: amf, smf, upf'
```
efiacor, line 223: Are quotes required here?

config/crd/bases/topology.nephio.org_nftopologies.yaml
s3wong: Yes --- this is the list of neighbor IDs, which is a simple way to record a set of connected NFs (according to how the template is instantiated). NFInstance is just a template --- every time a cluster is created with the specified labels, an instance of NF matching this template is deployed onto the cluster, and that instance forms a neighbor relationship on the existing NFs that aren't part of this NFInstance template but has attachment on the same network.

config/crd/bases/topology.nephio.org_nftopologies.yaml
s3wong: I removed this TODO on the update. I commented it out in this patch anyway, so just removing it would be fine.

config/crd/bases/topology.nephio.org_nftopologies.yaml
s3wong: the yaml is auto-generated from my comment on the go struct definition --- I don't have quotes in the comment. Probably just regenerate

s3wong:
Closing pull request as there is a need to refresh branch since the PR removes a file (nf_topology/v1alpha1/nf_topo_tracker.go) which caused conflict (i.e., not resolvable for now)

PR #56, opened 2023-10-25
Initial RAN CRD for R2

josephthaliath:
Related issue: https://github.com/nephio-project/nephio/issues/362


henderiw:
should this go under nfconfigs? we should align this with what sagar is doing and use a 3gpp.nephio group or so.
@johnbelamaric, @tliron  your view?

josephthaliath:
> should this go under nfconfigs? we should align this with what sagar is doing and use a 3gpp.nephio group or so. @johnbelamaric, @tliron your view?

@henderiw, Sagar's PR and this PR are using the same folder structures (nf_configs). I did not mention 3gpp, since we are not taking the parameters directly from there yet. It might lead to misunderstanding. Also for R2, it is very minimal list and this can be used for the R2 demonstration. I was thinking if SIG1 WG2 can work on a detail RAN CRD for R3?

henderiw:
https://github.com/nephio-project/api/pull/56#issuecomment-1780441563
we define a subset of 3gpp hence 3gpp.nephio.org or 3gpp.config.nephio.org should be used in my view. Lets have other people's view on this, but the current group is too specific

arora-sagar:
> [#56 (comment)](https://github.com/nephio-project/api/pull/56#issuecomment-1780441563) we define a subset of 3gpp hence 3gpp.nephio.org or 3gpp.config.nephio.org should be used in my view. Lets have other people's view on this, but the current group is too specific

Initially I did the same when we were discussing. But I wasn't sure if we can use the term 3GPP. Aren't there any license issues? If not then I don't mind changing it to `3gpp.nfconfig.nephio.org` from `nfconfig.nephio.org`

tliron:
> > [#56 (comment)](https://github.com/nephio-project/api/pull/56#issuecomment-1780441563) we define a subset of 3gpp hence 3gpp.nephio.org or 3gpp.config.nephio.org should be used in my view. Lets have other people's view on this, but the current group is too specific
> 
> Initially I did the same when we were discussing. But I wasn't sure if we can use the term 3GPP. Aren't there any license issues? If not then I don't mind changing it to `3gpp.nfconfig.nephio.org` from `nfconfig.nephio.org`

I'm in favor of keeping it 3gpp because that is most clear to users. This also helps us put a checkmark on SIG#1's intent for north-south-east-west integrations.

I think the trademark issue, if there is one, can be solved by adding a NOTICE file to repo (per the Apache License conventions). By the way, even using the terms "Kubernetes" and "K8s" should be noted there, as they are trademarks owned by CNCF. [Here's](https://github.com/tliron/turandot/blob/main/NOTICE) an example of how a NOTICE file can look.

I suggest raising topic in SIG#3 meeting and finding a general solution for us. Adding @radoslawc for his experience on licensing verification, perhaps we can add an automated test to make sure terms like "Kubernetes" and "3gpp" are accompanied by a NOTICE file.

gvbalaji:
> > > [#56 (comment)](https://github.com/nephio-project/api/pull/56#issuecomment-1780441563) we define a subset of 3gpp hence 3gpp.nephio.org or 3gpp.config.nephio.org should be used in my view. Lets have other people's view on this, but the current group is too specific
> > 
> > 
> > Initially I did the same when we were discussing. But I wasn't sure if we can use the term 3GPP. Aren't there any license issues? If not then I don't mind changing it to `3gpp.nfconfig.nephio.org` from `nfconfig.nephio.org`
> 
> I'm in favor of keeping it 3gpp because that is most clear to users. This also helps us put a checkmark on SIG#1's intent for north-south-east-west integrations.
> 
> I think the trademark issue, if there is one, can be solved by adding a NOTICE file to repo (per the Apache License conventions). By the way, even using the terms "Kubernetes" and "K8s" should be noted there, as they are trademarks owned by CNCF. [Here's](https://github.com/tliron/turandot/blob/main/NOTICE) an example of how a NOTICE file can look.
> 
> I suggest raising topic in SIG#3 meeting and finding a general solution for us. Adding @radoslawc for his experience on licensing verification, perhaps we can add an automated test to make sure terms like "Kubernetes" and "3gpp" are accompanied by a NOTICE file.

I am also not clear on this. On one hand we are not exactly mimicking 3gpp, Our models are based on them and also not tied to any particular version as well at this time.  We have not yet able to figure out the actual license model of 3gpp yang models.  As we evolve , we may deviate as well.   So I am not sure naming the group after 3gpp and thus inferring our models are compliant to 3gpp specs may send a wrong message. We can have the description / comments on the file saying the models are based on the 3gpp models.  What do you think?

henderiw:
> > > > [#56 (comment)](https://github.com/nephio-project/api/pull/56#issuecomment-1780441563) we define a subset of 3gpp hence 3gpp.nephio.org or 3gpp.config.nephio.org should be used in my view. Lets have other people's view on this, but the current group is too specific
> > > 
> > > 
> > > Initially I did the same when we were discussing. But I wasn't sure if we can use the term 3GPP. Aren't there any license issues? If not then I don't mind changing it to `3gpp.nfconfig.nephio.org` from `nfconfig.nephio.org`
> > 
> > 
> > I'm in favor of keeping it 3gpp because that is most clear to users. This also helps us put a checkmark on SIG#1's intent for north-south-east-west integrations.
> > I think the trademark issue, if there is one, can be solved by adding a NOTICE file to repo (per the Apache License conventions). By the way, even using the terms "Kubernetes" and "K8s" should be noted there, as they are trademarks owned by CNCF. [Here's](https://github.com/tliron/turandot/blob/main/NOTICE) an example of how a NOTICE file can look.
> > I suggest raising topic in SIG#3 meeting and finding a general solution for us. Adding @radoslawc for his experience on licensing verification, perhaps we can add an automated test to make sure terms like "Kubernetes" and "3gpp" are accompanied by a NOTICE file.
> 
> I am also not clear on this. On one hand we are not exactly mimicking 3gpp, Our models are based on them and also not tied to any particular version as well at this time. We have not yet able to figure out the actual license model of 3gpp yang models. As we evolve , we may deviate as well. So I am not sure naming the group after 3gpp and thus inferring our models are compliant to 3gpp specs may send a wrong message. We can have the description / comments on the file saying the models are based on the 3gpp models. What do you think?

I we do 3gpp.nephio.org it is actually 3gpp as seen by nephio. I don't see that we imply we use yang models of 3gpp. I would even say it does not imply we use yang at all. Now any other alternative is not easy:
- wireless -> wifi is also wireless and there are other technologies as well
- mobile -> wifi is also mobile and there are other technologies as well
This is why 3gpp is the best choice I see.

arora-sagar:
> > > > > [#56 (comment)](https://github.com/nephio-project/api/pull/56#issuecomment-1780441563) we define a subset of 3gpp hence 3gpp.nephio.org or 3gpp.config.nephio.org should be used in my view. Lets have other people's view on this, but the current group is too specific
> > > > 
> > > > 
> > > > Initially I did the same when we were discussing. But I wasn't sure if we can use the term 3GPP. Aren't there any license issues? If not then I don't mind changing it to `3gpp.nfconfig.nephio.org` from `nfconfig.nephio.org`
> > > 
> > > 
> > > I'm in favor of keeping it 3gpp because that is most clear to users. This also helps us put a checkmark on SIG#1's intent for north-south-east-west integrations.
> > > I think the trademark issue, if there is one, can be solved by adding a NOTICE file to repo (per the Apache License conventions). By the way, even using the terms "Kubernetes" and "K8s" should be noted there, as they are trademarks owned by CNCF. [Here's](https://github.com/tliron/turandot/blob/main/NOTICE) an example of how a NOTICE file can look.
> > > I suggest raising topic in SIG#3 meeting and finding a general solution for us. Adding @radoslawc for his experience on licensing verification, perhaps we can add an automated test to make sure terms like "Kubernetes" and "3gpp" are accompanied by a NOTICE file.
> > 
> > 
> > I am also not clear on this. On one hand we are not exactly mimicking 3gpp, Our models are based on them and also not tied to any particular version as well at this time. We have not yet able to figure out the actual license model of 3gpp yang models. As we evolve , we may deviate as well. So I am not sure naming the group after 3gpp and thus inferring our models are compliant to 3gpp specs may send a wrong message. We can have the description / comments on the file saying the models are based on the 3gpp models. What do you think?
> 
> I we do 3gpp.nephio.org it is actually 3gpp as seen by nephio. I don't see that we imply we use yang models of 3gpp. I would even say it does not imply we use yang at all. Now any other alternative is not easy:
> 
> * wireless -> wifi is also wireless and there are other technologies as well
> * mobile -> wifi is also mobile and there are other technologies as well
>   This is why 3gpp is the best choice I see.

@henderiw @tliron @gvbalaji @josephthaliath so as per today's meeting. Can I adapt the kind in my PR also to --> 
 `cellular.nephio.org` ?

henderiw:
so here are the groups we have:

config.nephio.org
workload.nephio.org
req.nephio.org
topology.nephio.org
ref.nephio.org

So is cellular.nephio.org not a bit odd ?

arora-sagar:
> so here are the groups we have:
> 
> config.nephio.org workload.nephio.org req.nephio.org topology.nephio.org ref.nephio.org
> 
> So is cellular.nephio.org not a bit odd ?

Sorry I saw later the minutes so everyone agreed on `cellular.config.nephio.org`  right?

johnbelamaric:
That is my understanding, yes.

arora-sagar:
> That is my understanding, yes.

Okay I did the changes in https://github.com/nephio-project/api/pull/57

josephthaliath:
> cellular.config.nephio.org

I have also updated the files according to this. If we get some more clarity, we can change it as required.

johnbelamaric (changes requested):
See my comments here: https://github.com/nephio-project/api/pull/57#pullrequestreview-1711123140

Same apply to this PR. 

The other thing I suggest is reviewing https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#api-conventions

One thing that probably applies to both this and #57 (@arora-sagar please take note) is the use of pointers for optional Go primitive types. At a glance I didn't see that done here. The reason is if you do not use a pointer, then Go will default it to the empty value for that type, which is not what you usually want - it makes it impossible to differentiate between "not set by the user" and "set to the empty value". Using a pointer works better from a defaulting standpoint, and is the convention for +optional fields. Note that maps and slices are already pointers so this is not necessary for those. But it is necessary for ints and strings (and of course structs).

josephthaliath:
> See my comments here: [#57 (review)](https://github.com/nephio-project/api/pull/57#pullrequestreview-1711123140)
> 
> Same apply to this PR.
> 
> The other thing I suggest is reviewing https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#api-conventions
> 
> One thing that probably applies to both this and #57 (@arora-sagar please take note) is the use of pointers for optional Go primitive types. At a glance I didn't see that done here. The reason is if you do not use a pointer, then Go will default it to the empty value for that type, which is not what you usually want - it makes it impossible to differentiate between "not set by the user" and "set to the empty value". Using a pointer works better from a defaulting standpoint, and is the convention for +optional fields. Note that maps and slices are already pointers so this is not necessary for those. But it is necessary for ints and strings (and of course structs).

Hi @johnbelamaric , I have made the changes according to the naming conventions.

PR #57, opened 2023-10-25
Common Configuration CRD for Core and RAN

arora-sagar:
**Description**
This PR addresses the common configuration CRD for Core and RAN. The common configurations I have verified will work for OAI 5G Core, Free 5GC, and Open5GS. I have created a CRD for AMF too, SMF and UPF don't need it for the moment but in the future, they will. Having a separate CRD for AMF allows fine-tuning AMF parameters. 

**Corresponding Issue**
https://github.com/nephio-project/nephio/issues/362 

**Note**
The verifications of the fields are based on how the standards specify. Some fields (`amfRegionID`,`amfPointer`,`amfSetID`) can be int and string both at the same time and they have different verification procedures. I didn't find a way to express it. 

This PR is related to https://github.com/nephio-project/api/pull/56 for adding RAN config CRDs.

**Dependency**
Once this PR is approved I can implement it in OAI 5G Core Network operators.


henderiw:
how would DNN work in this proposal? Does the den refer to the nfdeployment crd where the IP(s) resolve ?

arora-sagar:
> how would DNN work in this proposal? Does the den refer to the nfdeployment crd where the IP(s) resolve ?

@henderiw with this proposal DNN will be present in two places. So either we add PLMN/NSSAI/DNN combination in the same CRD as DNN or we remove from there and depend on this one. Or is there a way to link?

s3wong:
I actually like the `DNNInfo under S-NSSAI` piece as it concides with free5gc SMF config. But your DNNInfo is missing the UE pool, which is often associated with the DNN

nf_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+	// mcc defines the mobile country code
+	// +kubebuilder:validation:Pattern=`[02-79][0-9][0-9]`
+	Mcc string `json:"mcc" yaml:"mcc"`
+	// mnc defines the mobile network code
+	// +kubebuilder:validation:Pattern=`[0-9][0-9][0-9]|[0-9][0-9]`
+	Mnc string `json:"mnc" yaml:"mnc"`
+	// tac defines the tracking area code
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=16777215
+	Tac int `json:"tac" yaml:"tac"`
```
gvbalaji, line 64: NIt pick . You may want to have this as a pointer because if the value is not present marshaling will have 0 (zero value) instead of nil

nf_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+	// +optional
+	Nssai []Nssai `json:"nssai,omitempty" yaml:"nssai,omitempty"`
+}
+
+// NSSAI defines the Network Slice Selection Assistance Information
+type Nssai struct {
+	// Sst defines Slice/Service Type
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=255
+	Sst int `json:"sst" yaml:"sst"`
```
gvbalaji, line 75: NIt pick Same as above. You may need a pointer here.

nf_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+	PlmnInfo []PlmnInfoConfig `json:"plmnInfo,omitempty" yaml:"plmnInfo,omitempty"`
+	// communicationModel defines the communcation strategy between core components, using NRF or without NRF
+	// +optional
+	CommunicationModel []CommunicationModelConfig `json:"communicationModel,omitempty" yaml:"communicationModel,omitempty"`
+}
+
+// PlmnInfoConfig defines the structure of PLMN
+type PlmnInfoConfig struct {
+	// mcc defines the mobile country code
+	// +kubebuilder:validation:Pattern=`[02-79][0-9][0-9]`
```
gvbalaji, line 56: I know you mentioned the issue with ints and strings. Just curious looking at some of these, can they be straight ints?

arora-sagar:
> I actually like the `DNNInfo under S-NSSAI` piece as it concides with free5gc SMF config. But your DNNInfo is missing the UE pool, which is often associated with the DNN

@s3wong I did it on purpose to have this structure. But DNN pool is allocated in the DNN CR. I will not have it in advance. To fill in this structure. Unless I can have some linking and a controller fill that information in config CR.

nf_configs/v1alpha1/commoncoreranconfig_types.go
arora-sagar: @gvbalaji this value is mandatory. It should never be nil. It is the same for others too

nf_configs/v1alpha1/commoncoreranconfig_types.go
arora-sagar: @gvbalaji Not really because the standards say they can be both for example look here --> https://forge.3gpp.org/rep/sa5/MnS/-/blob/Rel-16/yang-models/_3gpp-common-yang-types.yang#L307

henderiw:
I believe the best solution is to put DNN in PLMN and build a specializer for this CR. So you create a package with the DNN requirement and put the result in the PLMN CR.

Now the challenge with this is you have to change the free5gc controller and someone need to create a specializer for this.
So the specialization loop would be:

specialization flow:
PLMN CR -> has DNN Req CR as is in the pkg -> DNN specializer requests IP (no change needed) -> IPAM specializer request IP (no change needed). -> a new specializer is required on PLMN CR which takes the DNN claimed IP and adds this in the PLMN CR.

In the controllers we need to use this config CR iso NF deployment to assign the PLMN/DNN and the related IP.

On top we need to update and take the DNN from the NF deployment CRD and move this into the PLMN CRD

arora-sagar:
> I believe the best solution is to put DNN in PLMN and build a specializer for this CR. So you create a package with the DNN requirement and put the result in the PLMN CR.
> 
> Now the challenge with this is you have to change the free5gc controller and someone need to create a specializer for this. So the specialization loop would be:
> 
> specialization flow: PLMN CR -> has DNN Req CR as is in the pkg -> DNN specializer requests IP (no change needed) -> IPAM specializer request IP (no change needed). -> a new specializer is required on PLMN CR which takes the DNN claimed IP and adds this in the PLMN CR.
> 
> In the controllers we need to use this config CR iso NF deployment to assign the PLMN/DNN and the related IP.
> 
> On top we need to update and take the DNN from the NF deployment CRD and move this into the PLMN CRD

The thing I don't understand is how to make a dependency in the PLMN CR for DNN, because in a PLMN there are multiple DNNs. So how should I reflect that. Can you make a small yaml to explain how to create this dependency? 

So far with the proposed CRD structure what I am doing is in SMF I have the dependency with UPF configuration (as it was earlier to get the DNN and ip-pool) and additional dependency to have this 3gpp.common-config CRD to be present. So I am fetching DNN information right now from two different places. I agree this is more error-prone.

To do it correctly I should fetch the information from only one place. If you can explain to me how to do it then let me see if I can write something in a short time or not.

arora-sagar:
@henderiw I have another idea (probably you proposed it too). How about I merge DNN and this CR, add the DNN pool here, and update the DNN specialized to search for the DNN pools from this CR? In that case the new CR would be something like this 
 
 
```
 apiVersion: nfconfig.nephio.org/v1alpha1
kind: CommonCoreRANConfig
metadata:
  name: common-core-ran-config
spec:
  plmnInfo:
    - mcc: "001"
      mnc: "01"
      tac: 0
      nssai:
        - sst: 1
          sd: ffffff
          dnnInfo:
            - name: internet
              sessionType: ipv4
              dns: 8.8.8.8
              pools:
                - name: pool1
                  prefixLength: 24
    - mcc: "208"
      mnc: "99"
      tac: 2
      nssai:
        - sst: 1
          sd: ffffff
          dnnInfo:
            - name: internet
              sessionType: ipv4
              dns: 8.8.8.8
  communicationModel: 
    - name: withNrf
      nrfUri: http://nrf-nnrf:8000 
```
What do you think?

henderiw:
The q is do we ever need a DNN outside of this CR or not? if not yes we can do this. If yes we should not and create another specializer acting upon the PLMN CR. 

So the idea is to make reusable building blocks we can reuse in various scenario's.

On top I also believe the DNS IP should be specialised and not be hardcoded.

The issue remains on the work split, are we able to pull this off for R2?

arora-sagar:
> The q is do we ever need a DNN outside of this CR or not? if not yes we can do this. If yes we should not and create another specializer acting upon the PLMN CR.
> 
> So the idea is to make reusable building blocks we can reuse in various scenario's.
> 
> On top I also believe the DNS IP should be specialised and not be hardcoded.
> 
> The issue remains on the work split, are we able to pull this off for R2?

We don't need DNN outside this CR, we don't need to have a special CR only for DNN. Yes true DNN Ip should not be hardcoded. It should be global DNS which is behind coredns.  

About the worksplit, I don't know. Who can do this?

nf_configs/v1alpha1/commoncoreranconfig_types.go
gvbalaji: ok thanks

nf_configs/v1alpha1/commoncoreranconfig_types.go
gvbalaji: ok thanks.

henderiw:
> > The q is do we ever need a DNN outside of this CR or not? if not yes we can do this. If yes we should not and create another specializer acting upon the PLMN CR.
> > So the idea is to make reusable building blocks we can reuse in various scenario's.
> > On top I also believe the DNS IP should be specialised and not be hardcoded.
> > The issue remains on the work split, are we able to pull this off for R2?
> 
> We don't need DNN outside this CR, we don't need to have a special CR only for DNN. Yes true DNN Ip should not be hardcoded. It should be global DNS which is behind coredns.
> 
> About the worksplit, I don't know. Who can do this?

Here are all the things that need to happen if we do this.
1. update DNN/NF deployment API
2. update DNN fn
3. Update the nf-deploy fn
4. Update the pipeline tests
5. update the packages of free5gc and later oai
6. update the free5gc controllers
7. Test E2E

arora-sagar:
> > > The q is do we ever need a DNN outside of this CR or not? if not yes we can do this. If yes we should not and create another specializer acting upon the PLMN CR.
> > > So the idea is to make reusable building blocks we can reuse in various scenario's.
> > > On top I also believe the DNS IP should be specialised and not be hardcoded.
> > > The issue remains on the work split, are we able to pull this off for R2?
> > 
> > 
> > We don't need DNN outside this CR, we don't need to have a special CR only for DNN. Yes true DNN Ip should not be hardcoded. It should be global DNS which is behind coredns.
> > About the worksplit, I don't know. Who can do this?
> 
> Here are all the things that need to happen if we do this.
> 
> 1. update DNN/NF deployment API
> 2. update DNN fn
> 3. Update the nf-deploy fn
> 4. Update the pipeline tests
> 5. update the packages of free5gc and later oai
> 6. update the free5gc controllers
> 7. Test E2E


That is true that this is quite some work and doing it in R2 will take some time. I have several options for R2. 

1. Take dnn from dnn CRD and keep this crd as it is. We will have two CRDs till R3
2. We postpone the integration of this CRD till R3

cellular_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CommonCoreRANStatus defines the observed state of CommonCoreRAN
+type CommonCoreRANConfigStatus struct {
+}
+
+// +kubebuilder:object:root=true
+type CommonCoreRANConfig struct {
```
henderiw, line 27: this name is not good. Is it not better to split this in 2 CRD(s)
1 for PLMN
1 for Communication Configuration.

It means we end up with potentially many CR(s), but it is composable.

other option is to keep It all in 1 CRD and use CommonConfig or just Config.
RANCore is very odd in my view

cellular_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+	// Sd defines Service Differentiator
+	// +optional
+	// +kubebuilder:validation:Pattern=`^[A-Fa-f0-9]{6}$`
+	Sd string `json:"sd,omitempty" yaml:"sd,omitempty"`
+	// dnnInfo defines the Data Network Names information
+	DnnInfo []DnnInfo `json:"dnnInfo,omitempty" yaml:"dnnInfo,omitempty"`
+}
+
+// DNN defines the Data Network Names Information
+type DnnInfo struct {
```
henderiw, line 85: so this is a reference and the name need to match the NF Deployment. Right?
If so make this clear and add a NOTE. 
One R2 is out we have to refactor this.

config/crd/bases/cellular.config.nephio.org_amfconfigs.yaml
```diff
@@ -0,0 +1,101 @@
...
+                items:
+                  description: ServedGuami information related to GUAMI
+                  properties:
+                    amfId:
+                      description: AmfId defines the AMF Identifier
+                      properties:
+                        amfPointer:
+                          description: amfPointer uniquely identifies the AMF in AMF
+                            set
+                          maxLength: 6
```
efiacor, line 53: Does the lenght need to be restricted for these?
Is this from the 3pp spec?

cellular_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+	Items           []CommonCoreRANConfig `json:"items" yaml:"items"`
+}
+
+// CommonCoreRANConfigSpec defines the characteristics of a deployment of a network function
+type CommonCoreRANConfigSpec struct {
+	// plmnInfo defines the list of PLMNs to configure for core NFs
+	PlmnInfo []PlmnInfoConfig `json:"plmnInfo,omitempty" yaml:"plmnInfo,omitempty"`
+	// communicationModel defines the communcation strategy between core components, using NRF or without NRF
+	// +optional
+	CommunicationModel []CommunicationModelConfig `json:"communicationModel,omitempty" yaml:"communicationModel,omitempty"`
```
henderiw, line 50: When and why do we have multiple communication models? What is the scenario ?

cellular_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// PlmnInfoConfig defines the structure of PLMN
+type PlmnInfoConfig struct {
```
henderiw, line 54: just clarifying that all these parameters are mandatory except for NSSAI ?

cellular_configs/v1alpha1/amfconfig_types.go
```diff
@@ -0,0 +1,86 @@
...
+	metav1.ListMeta `json:"metadata,omitempty" yaml:"metadata,omitempty"`
+	Items           []AmfConfig `json:"items" yaml:"items"`
+}
+
+// AmfConfigSpec defines the characteristics of a deployment of a network function
+type AmfConfigSpec struct {
+	// amfName defines name of the amf NF
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	AmfName string `json:"amfName" yaml:"amfName"`
```
henderiw, line 49: can the name not come from the CR ?

config/crd/bases/cellular.config.nephio.org_commoncoreranconfigs.yaml
efiacor, line 1: Being pedantic here but could we add under scores to the name for readability?

cellular_configs/v1alpha1/amfconfig_types.go
```diff
@@ -0,0 +1,86 @@
...
+	// mcc defines the mobile country code
+	// +kubebuilder:validation:Pattern=`[02-79][0-9][0-9]`
+	Mcc string `json:"mcc" yaml:"mcc"`
+	// mnc defines the mobile network code
+	// +kubebuilder:validation:Pattern=`[0-9][0-9][0-9]|[0-9][0-9]`
+	Mnc string `json:"mnc" yaml:"mnc"`
+}
+
+// AmfId defines the AMF Identifier
+type AmfId struct {
```
henderiw, line 73: are all these parameters mandatory ?

cellular_configs/v1alpha1/amfconfig_types.go
```diff
@@ -0,0 +1,86 @@
...
+// ServedGuami information related to GUAMI
+type ServedGuamiConfig struct {
+	// PlmnId defines the PLMN Identifier
+	PlmnId PlmnId `json:"plmnId" yaml:"plmnId"`
+	// AmfId defines the AMF Identifier
+	AmfId AmfId `json:"amfId" yaml:"amfId"`
+}
+
+// PlmnId defines the Public Land Mobile Network Identifier
+type PlmnId struct {
```
henderiw, line 63: why can we not reuse this from common config  or vice versa?
This is duplication.

henderiw: we can use json:inline for this to inherit the json tags.

cellular_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=16777215
+	Tac int `json:"tac" yaml:"tac"`
+	// NSSAI defines the Network Slice Selection Assistance Information
+	// +optional
+	Nssai []Nssai `json:"nssai,omitempty" yaml:"nssai,omitempty"`
+}
+
+// NSSAI defines the Network Slice Selection Assistance Information
+type Nssai struct {
```
henderiw, line 71: are all these parameters mandatory

henderiw (changes requested):

cellular_configs/v1alpha1/amfconfig_types.go
```diff
@@ -0,0 +1,86 @@
...
+limitations under the License.
+*/
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// AmfConfigStatus defines the observed state of AmfConfig
+type AmfConfigStatus struct {
```
johnbelamaric, line 23: Convention is to match the case of the entire "initialism" to the case of the first character. So, in struct names "Amf" -> "AMF". In exported Go field names the same. In the JSON names, "amf" if it's the start of the field name, but "AMF" if it appears after the first word.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions

That is, you need to change "Amf" to either "amf" or "AMF" everywhere, in every file, based upon where it fits in the convention.

cellular_configs/v1alpha1/amfconfig_types.go
johnbelamaric: agreed, is there a reason this should not match metadata.name?

cellular_configs/v1alpha1/amfconfig_types.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// AmfConfigSpec defines the characteristics of a deployment of a network function
+type AmfConfigSpec struct {
+	// amfName defines name of the amf NF
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	AmfName string `json:"amfName" yaml:"amfName"`
+	// ServedGuami defines information related to GUAMI
+	ServedGuami []ServedGuamiConfig `json:"servedGuami" yaml:"servedGuami"`
```
johnbelamaric, line 51: This should be "ServedGUAMI" and "servedGUAMI" per conventions.

johnbelamaric (changes requested):
Reviewing really from a K8s convention perspective, the main issue I see is that all the various initialism (AMF, GUAMI, MCC, PLMN, etc.) should not mix case. 

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions

johnbelamaric:
@arora-sagar please also take a look through this in light of this comment regarding pointers for optional fields: https://github.com/nephio-project/api/pull/56#pullrequestreview-1711140048

cellular_configs/v1alpha1/amfconfig_types.go
s3wong: PLMN is also needed for free5gc SMF --- so it seems to be something that shouldn't be specific to AMF

cellular_configs/v1alpha1/amfconfig_types.go
s3wong: Certainly not needed for free5gc

cellular_configs/v1alpha1/commoncoreranconfig_types.go
s3wong: I think for OAI, UPF talks to NRF, whereas in free5gc, UPF doesn't even know what a NRF is

cellular_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+	// name defines dnn
+	Name string `json:"name" yaml:"name"`
+	// SessionType defines session type
+	// +kubebuilder:validation:Enum=ipv4;ipv6;ipv4v6;unstructured;ethernet
+	SessionType string `json:"sessionType" yaml:"sessionType"`
+	// Dns defines Domain Network Service Name
+	// +optional
+	Dns string `json:"dns" yaml:"dns"`
+}
+
```
s3wong, line 95: We need to associate a data network with a UE IP pool --- I don't see where we would get that information from this definition...

cellular_configs/v1alpha1/amfconfig_types.go
arora-sagar: Okay I will make this change, thanks for the conventions.

cellular_configs/v1alpha1/amfconfig_types.go
arora-sagar: True there isn't any specific reason, I can remove it. Also metadata. name I suppose can be `253` characters so it should be fine.

cellular_configs/v1alpha1/amfconfig_types.go
arora-sagar: Yes, I will adjust all the param according to the conventions.

cellular_configs/v1alpha1/amfconfig_types.go
arora-sagar: @henderiw This is not really duplication, how should I mention the `amdId` under the corresponding `plmnId` if I take the `plmnId` from the common crd? Also here it is GUAMI --> Globally Unique AMF ID

@s3wong in SMF we need only plmnID

```
plmnList:
  - mcc: "208"
    mnc: "93"
```

cellular_configs/v1alpha1/amfconfig_types.go
arora-sagar: @henderiw @s3wong I made them mandatory according to yang models, I can remove them from mandatory. Also `amfId` is made using these 3 values, free5GC combines them to express `amfId`.

cellular_configs/v1alpha1/commoncoreranconfig_types.go
arora-sagar: @henderiw I called it CommonCoreRANConfig because the PLMN/NSSAI part is common for RAN also I took the parameters that are common among all the core NFs. 

Right now I only proposed CRD specific to AMF, but later we can have specific ones for SMF, UPF and other NFs if needed. 

But If you want I can change it to `CommonCoreConfig` and let the RAN to have its own parameters. 

@josephthaliath do you have an opinion here?

cellular_configs/v1alpha1/commoncoreranconfig_types.go
arora-sagar: @henderiw @s3wong not really again this comes from 3gpp yang models. In OAI also I can do something similar to free5GC. Communication model depicts if NFs communicate with NRF (5G service based architecture) or not (5G reference based architecture). 

If Nephio will only deploy 5G Service Based Architecture then I can remove this parameter. I just wanted to have a way to deploy both architectures if possible and if needed by someone.

cellular_configs/v1alpha1/commoncoreranconfig_types.go
arora-sagar: Okay I will add that @henderiw thanks

cellular_configs/v1alpha1/commoncoreranconfig_types.go
arora-sagar: @s3wong The UE-pool information for R2 we need to get it from `DNN` CR because there is no specializer for now that will fetch the information from DNN CR and mention it here. 

@henderiw this is what was proposed right? Only for R2?

config/crd/bases/cellular.config.nephio.org_amfconfigs.yaml
arora-sagar: Yes @efiacor I took from yang models and specs the limit all the limits

config/crd/bases/cellular.config.nephio.org_commoncoreranconfigs.yaml
arora-sagar: All of them? Or specific ones? Can you give me an example?

nf_configs/v1alpha1/commoncoreranconfig_types.go
```diff
@@ -0,0 +1,103 @@
...
+	Items           []CommonCoreRANConfig `json:"items" yaml:"items"`
+}
+
+// CommonCoreRANConfigSpec defines the characteristics of a deployment of a network function
+type CommonCoreRANConfigSpec struct {
+	// plmnInfo defines the list of PLMNs to configure for core NFs
+	PlmnInfo []PlmnInfoConfig `json:"plmnInfo,omitempty" yaml:"plmnInfo,omitempty"`
+	// communicationModel defines the communcation strategy between core components, using NRF or without NRF
+	// +optional
+	CommunicationModel []CommunicationModelConfig `json:"communicationModel,omitempty" yaml:"communicationModel,omitempty"`
```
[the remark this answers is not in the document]
henderiw: I have been thinking about this and I believe we should do an aggregate CRD that encapsulates smaller entities that are contained by themselves. As such we avoid adding these CRDs to the API server -> and having only the aggregate applied to the cluster. This is similar how the reference CRD works.

cellular_configs/v1alpha1/amfconfig_types.go
henderiw: if you use json:inline you don't need to redefine these parameters.

cellular_configs/v1alpha1/commoncoreranconfig_types.go
henderiw: yes for R2 the name of the inn here should match the one in the nf deployment. It is clunky but it avoids doing all the other work required. So for R2 we use this hack for now.

cellular_configs/v1alpha1/commoncoreranconfig_types.go
josephthaliath: @arora-sagar, @henderiw , If we think only PLMN Info is common between RAN and Core , then we can create a seperate CRD for this and name it PLMNInfo itself. But if we think we will add more parameters  in future which are common between RAN and Core, then a CRD with name CommonConfig can be used as placeholder for such parameters. Domething like this:
1) CommonConfig ( common across RAN and Core  NFs) 
2) CommonCoreConfig ( common across Core NFs )
3) CommonRANConfig (common across all RAN NFs )

nf_configs/v1alpha1/commoncoreranconfig_types.go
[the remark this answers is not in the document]
josephthaliath: @henderiw , I was assuming we are taking the dependency approach and these config CRs are referred in the dependency yamls in the package. In this approach finally config CR with embedded config CRs are applied. Is his the right understanding?

cellular_configs/v1alpha1/commoncoreranconfig_types.go
henderiw: My point is for a given deployment do you need more then 1 ?

cellular_configs/v1alpha1/commoncoreranconfig_types.go
henderiw: So I am of the opinion we need only 1 CRD that embeds the other smaller CRD that define a unit of x (e.g. PLMN is such a small unit).

So to depict it we call this config.cellular.config.nephio.org. it is the meta CR that is applied to the cluster and embeds these smaller self contained elements. As such we don't need to care about RAN or Core. We apply it to the NF deployment through the parameterRef. So no need for RAN or Core. Just it is applied to the NF of choice. You parse the CR units from there after you picked up the CR.

Now the only issue might be CR size but we can solve this by adding multiple configs with a different name.
Thoughts ?

cellular_configs/v1alpha1/commoncoreranconfig_types.go
arora-sagar: @henderiw For a given deployment there is only one communication model either 
with NRF or without NRF.

nf_configs/v1alpha1/commoncoreranconfig_types.go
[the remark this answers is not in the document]
henderiw: Here is how the upper config CRD would look like:
It references the other CR that are embedded in the uber CR.

package v1alpha1

import (
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/runtime"
)

// +kubebuilder:object:root=true
// +kubebuilder:subresource:status
type Config struct {
	metav1.TypeMeta   `json:",inline" yaml:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty" yaml:"metadata,omitempty"`

	Spec   ConfigSpec   `json:"spec,omitempty" yaml:"spec,omitempty"`
	Status ConfigStatus `json:"status,omitempty" yaml:"status,omitempty"`
}

type ConfigSpec struct {
	// Config is the embedded config
	//+kubebuilder:pruning:PreserveUnknownFields
	ConfigRefs []runtime.RawExtension `json:"config,omitempty" yaml:"config,omitempty"`
}

type ConfigStatus struct {
}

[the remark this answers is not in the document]
josephthaliath: @henderiw , yes this my understanding as well. When you mentioned applied in the cluster, I was thinking it is applied as separate resource and not embedded in the Config. Now it is clear. Thanks

cellular_configs/v1alpha1/amfconfig_types.go
```diff
@@ -19,68 +19,58 @@ import (
...
 
-// AmfConfigSpec defines the characteristics of a deployment of a network function
-type AmfConfigSpec struct {
+// AMFConfigSpec defines the characteristics of a deployment of a network function
+type AMFConfigSpec struct {
 	// amfName defines name of the amf NF
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
-	AmfName string `json:"amfName" yaml:"amfName"`
+	AMFName string `json:"amfName" yaml:"amfName"`
```
henderiw, line 49: I thought we were going to remove this

cellular_configs/v1alpha1/amfconfig_types.go
```diff
@@ -19,68 +19,58 @@ import (
...
 
-// AmfConfigStatus defines the observed state of AmfConfig
-type AmfConfigStatus struct {
+// AMFConfigStatus defines the observed state of AmfConfig
+type AMFConfigStatus struct {
 }
 
 // +kubebuilder:object:root=true
-type AmfConfig struct {
+type AMFConfig struct {
```
henderiw, line 27: Do we call this AMF as this is a subset of the AMF. Seems it is better called ServedGuami

cellular_configs/v1alpha1/config_types.go
```diff
@@ -0,0 +1,40 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty" yaml:"metadata,omitempty"`
+
+	Spec   ConfigSpec   `json:"spec,omitempty" yaml:"spec,omitempty"`
+	Status ConfigStatus `json:"status,omitempty" yaml:"status,omitempty"`
+}
+
+type ConfigSpec struct {
+	// Config is the embedded config
+	//+kubebuilder:pruning:PreserveUnknownFields
+	ConfigRefs []runtime.RawExtension `json:"config,omitempty" yaml:"config,omitempty"`
```
henderiw, line 39: make the name of the tag the same as the Name of the Param. Should configRef

cellular_configs/v1alpha1/plmnconfig.go
```diff
@@ -19,68 +19,71 @@ import (
...
 
-// CommonCoreRANStatus defines the observed state of CommonCoreRAN
-type CommonCoreRANConfigStatus struct {
+// PLMNConfigStatus defines the observed state of PLMNConfig
+type PLMNConfigStatus struct {
 }
 
 // +kubebuilder:object:root=true
-type CommonCoreRANConfig struct {
+type PLMNConfig struct {
```
henderiw, line 27: call this PLMN

henderiw (changes requested):

cellular_configs/v1alpha1/plmnconfig.go
```diff
@@ -0,0 +1,97 @@
...
+// NSSAI defines the Network Slice Selection Assistance Information
+type NSSAI struct {
+	// Sst defines Slice/Service Type
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=255
+	SST int `json:"sst" yaml:"sst"`
+	// Sd defines Service Differentiator
+	// +optional
+	// +kubebuilder:validation:Pattern=`^[A-Fa-f0-9]{6}$`
+	SD string `json:"sd,omitempty" yaml:"sd,omitempty"`
```
johnbelamaric, line 81: Optional strings should be a pointer to a string.

cellular_configs/v1alpha1/plmnconfig.go
```diff
@@ -0,0 +1,97 @@
...
+	// dnnInfo defines the Data Network Names information
+	DNNInfo DnnInfo `json:"dnnInfo" yaml:"dnnInfo"`
+}
+
+// NOTE: For R2 the dnn.name here should match with NF deployment dnn name. This is a hack for R2
+// DNN defines the Data Network Names Information
+type DnnInfo struct {
```
johnbelamaric, line 88: s/Dnn/DNN/g

cellular_configs/v1alpha1/plmnconfig.go
```diff
@@ -0,0 +1,97 @@
...
+	// name defines dnn
+	Name string `json:"name" yaml:"name"`
+	// SessionType defines session type
+	// +kubebuilder:validation:Enum=ipv4;ipv6;ipv4v6;unstructured;ethernet
+	SessionType string `json:"sessionType" yaml:"sessionType"`
+	// Dns defines Domain Network Service Name
+	// +optional
+	DNS string `json:"dns" yaml:"dns"`
```
johnbelamaric, line 96: What is a "Domain Network Service Name"? Is it different than a "Domain Name Service", which is the usual meaning of "DNS" as an initialism. If it's different, please don't use "DNS".

Or, is this a domain name? If so, Can you just call it DomainName or Zone or ZoneName or DNSZone? "DNS" makes me wonder if it is, for example, the DNS server IP or something.

If it's a zone name it needs to be RFC compliant, there are existing k8s validations for this.

And optional strings should be pointers.

cellular_configs/v1alpha1/amfconfig_types.go
arora-sagar: I just kept ServedGUAMI for the moment but later we can add more so it is better to call it AMF to allow expansion.

cellular_configs/v1alpha1/plmnconfig.go
arora-sagar: @johnbelamaric My bad it is DNS that the User Equipment uses. So here we have to specify the IP address of the DNS.

cellular_configs/v1alpha1/plmnconfig.go
arora-sagar: Done

cellular_configs/v1alpha1/plmnconfig.go
arora-sagar: Changed

cellular_configs/v1alpha1/config_types.go
arora-sagar: Done

cellular_configs/v1alpha1/amfconfig_types.go
arora-sagar: Done

cellular_configs/v1alpha1/config_types.go
arora-sagar: In fact, I made it `configs` now for both, I think it is better than configRef to avoid any confusion

arora-sagar:
@henderiw I corrected to kinds to

1. plmns.cellular.config.nephio.org
2. amfs.cellular.config.nephio.org (Though there is only servedGUAMI for the moment in the future we can add more fields else there will be a lot of small CRs).  If you are really in favor of only continuing with small CRs then okay I will keep it servedGuami. 
3.configs.cellular.config.nephio.org (I made the raw field configs rather than configRef (that will create confusion I suppose)

When you have time can you give me an example of how a package should be made? 

@johnbelamaric I hope now the naming is correct according to the convention I checked K8s apis too.

cellular_configs/v1alpha1/plmn_types.go
```diff
@@ -0,0 +1,97 @@
...
+// DNN defines the Data Network Names Information
+type DNNInfo struct {
+	// name defines DNN
+	Name string `json:"name" yaml:"name"`
+	// SessionType defines session type
+	// +kubebuilder:validation:Enum=ipv4;ipv6;ipv4v6;unstructured;ethernet
+	SessionType string `json:"sessionType" yaml:"sessionType"`
+	// DNS defines Domain Name Server
+	// +optional
+	DNS *string `json:"dns" yaml:"dns"`
```
johnbelamaric, line 96: I am still not clear on what this is. Is it the resolver address the CPE should use? Is it an IP address? If so, maybe "NameServerIPAddress"? Can there be more than one? Then it should be plural.

cellular_configs/v1alpha1/served_guami_types.go
```diff
@@ -0,0 +1,66 @@
...
+	// AMFID defines the AMF Identifier
+	AMFID AMFID `json:"amfID" yaml:"amfID"`
+}
+
+// AMFID defines the AMF Identifier
+type AMFID struct {
+	// amfRegionId identifies the region
+	// +kubebuilder:validation:Pattern=`[01]*`
+	// +kubebuilder:validation:MaxLength=8
+	AMFRegionId string `json:"amfRegionID" yaml:"amfRegionID"`
```
johnbelamaric, line 57: Should these be "ID" not "Id"? Are they required?

cellular_configs/v1alpha1/plmn_types.go
arora-sagar: @johnbelamaric It is an IP Address and what you understand is correct (NameServerIPAddress). I used the term DNS because most of the configuration files (Free5GC, OAI, Open5GS Core Network) use it as `dns`.

I have attached the screenshot of the field from the Wireshark packet capture when CPE/UE has a successful session towards a data network. These information elements are standardized 

image

config/crd/bases/cellular.config.nephio.org_configs.yaml
```diff
@@ -0,0 +1,49 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.13.0
+  name: configs.cellular.config.nephio.org
+spec:
+  group: cellular.config.nephio.org
```
arora-sagar, line 9: Change group to --> `nf.nephio.org` and kind to `Config` according to the discussion with @henderiw and @josephthaliath because it allows extending to other type of network functions except cellular network functions.

arora-sagar:
Closing this PR as decided with @henderiw and @josephthaliath this approach will be changed in R3 with yang translators. For R2 the small CRs should be part of the experimental repository and config meta CRD of the Api repository. I will open another PR with only meta CRD.

henderiw:
do you open a new PR for the generic config one?

PR #59, opened 2023-11-24
NFConfig Meta CRD

arora-sagar:
**Description**

Created NFConfig Meta CRD (Similar to ConfigRef CRD) to contain configuration. 

**Breaking change**

I had to change the `nf_deployments` folder to workload because the naming wasn't chosen as per the group. So we need to adjust the free5GC tutorial before accepting this PR.


henderiw:
We have to do it in stages.For a dir change the free5gc does not have to change as we semantically did not change anything on NFdeployment. Now it is always better keep the repos in sync but it is not mandatory in this case.

PR #60, opened 2023-11-27
(bug): adding missing fn

arora-sagar:
In the previous commit, `DeepCopyObject` was not added. I didn't mention `//+kubebuilder:object:root=true` on top of NFConfigList.


arora-sagar:
@josephthaliath Can you check if this is okay?

PR #61, opened 2023-12-13
Simplified NFTopology CRDs

s3wong:


nf_topology/v1alpha1/nf_topology_types.go
```diff
@@ -41,13 +32,30 @@ type NFInterface struct {
...
+
+	// Revision is the specific version number of the revision of the package
+	Revision string `json:"revision"`
+}
+
 // NFTemplate defines the template for deployment of an instance of a NF
 type NFTemplate struct {
 	// NFType specifies the type of NF this template is specifying
-	NFType NFType `json:"nfType" yaml:"nfType"`
+	NFType string `json:"nfType" yaml:"nfType"`
```
liamfallon, line 55: Yes, better not to use an enum to avoid having to identify all possible types of NFs.

liamfallon:
Let's close this for now and we can come back to it later.

PR #63, opened 2024-12-19
Adding O2IMS CRD for create cluster story

dkosteck:
This O2IMS CRD is defined for the northbound API of Nephio, and data will be given by FOCOM. Therefore we should respect the data that FOCOM will provide, and in Specs a Provisioning Request is defined and there is no notion of namespaces

### DIFF
--- a/config/crd/bases/o2ims.provisioning.oran.org_provisioningrequests.yaml
+++ b/config/crd/bases/o2ims.provisioning.oran.org_provisioningrequests.yaml
@@ -162,69 +162,6 @@ spec:
                 description: |-
                   Extensions contain extra details about the resources and the configuration used for/by
                   the ProvisioningRequest.
-                properties:
-                  clusterDetails:
-                    description: ClusterDetails references to the ClusterInstance.
-                    properties:
-                      clusterProvisionStartedAt:
-                        description: A timestamp indicating the cluster provisoning
-                          has started
-                        format: date-time
-                        type: string
-                      name:
-                        description: Contains the name of the created ClusterInstance.
-                        type: string
-                      nonCompliantAt:
-                        description: Holds the first timestamp when the configuration
-                          was found NonCompliant for the cluster.
-                        format: date-time
-                        type: string
-                      ztpStatus:
-                        description: Says if ZTP has complete or not.
-                        type: string
-                    type: object
-                  nodePoolRef:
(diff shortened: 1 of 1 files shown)


henderiw:
this is not a good api definition. You dont call something a provisionRequest. K8s apis are about the what not the how.
also the scope cluster is not good in my view as this constraints things a lot for no reason. this is about requesting clusters again.

arora-sagar:
> this is not a good api definition. You dont call something a provisionRequest. K8s apis are about the what not the how. also the scope cluster is not good in my view as this constraints things a lot for no reason. this is about requesting clusters again.

@henderiw I understand the term "Requests" is not suitable. But it comes from O-RAN cluster provisioningRequest. This CRD was accepted in the last WG2 (O-RAN meeting). Now, we need to discuss this in the meeting to change it. 

I am fine in changing from cluster scope to namespace

arora-sagar:
@gvbalaji Do you have some comments?

gvbalaji:
Yes as @arora-sagar mentioned this is the CRD from O-RAN specs we're using as is and adapting to our internal implementation. This is used for our thin IMS implementation.

henderiw:
ORAN should follow the k8s principles. it is not because it is approved there we have to accept it. You don't call a CRD kind using verbs. A k8s api/crd should be about what the resource is about. it should be Cluster or ClusterClaim and given you have a oran group it is relevant in the oran context.

gupta-alok:
I agree with @arora-sagar and @gvbalaji. The scope and objective of this PR is to implement O-RAN defined constructs and APIs. We in Nephio shall not change (rename) O-RAN defined IOCs such as ProvisioningRequest for lifecycle management of O-Cloud Node Clusters as well as O-Cloud Infrastructure resource(s).

Nephio-based O-Cloud IMS implementation will have its internal Nephio CRs such as **ClusterClaim** and **WorkloadCluster** CRs.

The **ClusterClaim** CR will model the request to create a cluster with a certain configuration while the current **WorkloadCluster** CR will be used to model successfully deployed clusters and used as a target for NF Deployment. The design details regarding Create O-Cloud Node Cluster User Story can be found here: https://docs.google.com/presentation/d/1iyE_XLkBJkl_ClYsE_vRnEkm8Ica7YE0/edit#slide=id.p1

PR #65, opened 2025-01-24
Update O2IMS Provisioning Request CRD based on updates from O-RAN CR

dkosteck:
The main change to the CRD to the status block and conditions fields, this is from the O-RAN.WG6.TS.O-CLOUD-IM.0-R004-v03.00 CR rev 00.03


liamfallon (commented):
/approve

PR #66, opened 2025-02-03
Update O2IMS provisionedResources to provisionedResourceSet per CRD

dkosteck:
Alok pointed out that the original yaml we received had an old field that did not match the CR, updating that field.

### DIFF
--- a/config/crd/bases/o2ims.provisioning.oran.org_provisioningrequests.yaml
+++ b/config/crd/bases/o2ims.provisioning.oran.org_provisioningrequests.yaml
@@ -165,7 +165,7 @@ spec:
                 type: object
               provisioningStatus:
                 properties:
-                  provisionedResources:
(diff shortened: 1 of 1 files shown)


efiacor:
@dkosteck requires a rebase I believe.

dkosteck:
> @dkosteck requires a rebase I believe.

@efiacor Yes, looks like it. Let's wait to see if we get a thumbs up from Alok or Stefan on the CRD since they pointed out the issue with the naming, and originally supplied the updated CRD.

gupta-alok:
LGTM. Thank you.

liamfallon:
@dkosteck I think we can go ahead with the rebase now if you like.

dkosteck:
@liamfallon @efiacor Just merged the machines from main into this branch, should now have no conflicts

